PR #95, opened 2022-01-20
add mariadb clients

rolandknight:
add mariadb clients (mariadb and mariadb-dump)


alecthomas:
Are there no Mac packages available?

rolandknight:
Official Mac packages aren't available anymore:
https://mariadb.com/kb/en/mariadb-pkg/#comment_3703

alecthomas:
Sorry for the delay, I was pondering whether we want to do this, but I think it's okay.

mariadb-clients.hcl
```diff
@@ -0,0 +1,10 @@
+description = "Mariadb client."
```
alecthomas, line 1: Can you please rename it to `mariadb-client.hcl`, so it's consistent with the existing `mysql-client.hcl`.

alecthomas:
Thanks!

PR #102, opened 2022-02-10
xcodegen 2.26.0

quad:
xcodegen normally installs via homebrew and puts files in `/usr/local/share/xcodegen/SettingPresets`. I'm not sure if this absence will affect a Hermit installation.

Consider this a WIP as I test this on my project.


quad:
It appears to work!

PR #105, opened 2022-02-19
Tanka

rolandknight:
tanka kubernetes installer (similar to helm)


tanka.hcl
```diff
@@ -0,0 +1,31 @@
+description = "Flexible, reusable and concise configuration for Kubernetes."
+binaries = ["tk"]
+test = "tk --help"
+
+linux {
```
alecthomas, line 5: FYI if you have two platform blocks that are identical, as these two are, you can just pull them out to the top level.

alecthomas:
Can you please clean up your fork so that commits don't keep accumulating?

alecthomas:
Thanks!

rolandknight:
Done. Created another clean pull request that simplifies tanka and adds
jsonnet-bundler (it's a tool that tanka uses BTW)

PR #115, opened 2022-03-16
meson 0.16.3

quad:


quad:
@jmecom you may also find this of interest.

PR #197, opened 2022-09-07
feat: add dotnet versions

stillmatic:
the JQ parsing to get to a reasonable file format is not bad, but needed some python to format the json output in the expected hcl format


alecthomas (approved):
Oof, curse you Microsoft!

PR #198, opened 2022-09-08
feat: add the mysqladmin binary as part of mysql-client package

staktrace:


staktrace:
Hm looks like on linux it's in a separate tarball. I'll make a new PR

PR #212, opened 2022-10-06
feat(velero.hcl): add Velero manifest

usrme:


usrme:
Interesting... I was able to run `hermit test velero --trace` successfully prior to pushing this :thinking:

PR #266, opened 2023-01-10
Add pip/python to python3

damienrj:
This should avoid the earlier problems by no longer referencing specific binaries. The cache needs to be cleared to see the new python/pip aliases.

### DIFF
--- a/python3.hcl
+++ b/python3.hcl
@@ -1,6 +1,6 @@
 description = "Python is a programming language that lets you work quickly and integrate systems more effectively."
 strip = 1
-binaries = ["install/bin/pip3", "install/bin/python3"]
+binaries = ["install/bin/p*"]
 test = "python3 -m pip install flake8"
 env = {
   "PYTHONPYCACHEPREFIX": "${HERMIT_ENV}/.hermit/python/cache",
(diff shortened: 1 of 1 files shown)


python3.hcl
```diff
@@ -1,35 +1,26 @@
...
 test = "python3 -m pip install flake8"
+
 env = {
-  "PYTHONPYCACHEPREFIX": "${HERMIT_ENV}/.hermit/python/cache",
-  "PIP_CACHE_DIR": "${HERMIT_ENV}/.hermit/python/cache/pip",
-  "PYTHONUSERBASE": "${HERMIT_ENV}/.hermit/python",
-  "PATH": "${HERMIT_ENV}/.hermit/python/bin:${PATH}",
+  PYTHONPYCACHEPREFIX: "${HERMIT_ENV}/.hermit/python/cache",
+  PYTHONUSERBASE: "${HERMIT_ENV}/.hermit/python",
+  PATH: "${HERMIT_ENV}/.hermit/python/bin:${PATH}",
```
alecthomas, line 9: This has dropped an envar, PIP_CACHE_DIR

alecthomas:
Have you tested this with the older packages?

damienrj:
I tested this with a package using an older python version and it didn't
have any issues I believe that it wasn't specifying explicitly the pip and
python binaries. I had to clear the cache so that the python would unpack
and make the links though.

python3.hcl
damienrj: Ended up reverting the formatting

alecthomas (approved):

PR #335, opened 2023-06-01
Adds zulu8.70.0.23 for support for m1 macs working in java 1.8.

cassiechew:
This build seems to work better, having a build for m1 arch.
Good for older systems still using JDK 8 but using m1 macs.


jvmakine (approved):

cassiechew:
Thanks so much

PR #339, opened 2023-06-04
feat: add dragonflydb

stillmatic:
https://github.com/dragonflydb/dragonfly

note: does not support macos. what is the right way to handle that? https://www.dragonflydb.io/docs/getting-started/binary#prerequisites


alecthomas (approved):
Hermit will warn the user when they're installing.

PR #367, opened 2023-08-10
chore: add versions `1.15.4`, `1.16.2`, & `1.17.1` for flink using the new dlcnd sub-domain.

MatthewDolan:
According to Apache Downloads (https://www.apache.org/dyn/closer.cgi):

> We suggest the following site for your download:
>
> https://dlcdn.apache.org/

It is my understanding that the archive was previously used and kept for historical reasons.


MatthewDolan (commented):
Actually, on second look, I discovered that there are two subdomains (https://downloads.apache.org and https://dlcdn.apache.org/) and I am not sure which is preferable.

alecthomas (approved):

PR #379, opened 2023-09-06
Adds version 0.6.2 of the codecov uploader

RONNCC:
Adds version 0.6.2 of the codecov uploader since codecov support indicated we're running into bugs since we're using an old uploader version in this ticket: https://codecoventerprise.codecov.io/hc/en-us/requests/102089

image

where CodeCov is counting comments as part of it's code quality inspections


lyonlai (approved):
might be worth looking at make auto-version working for this package so that you don't have to do it manuallly

PR #418, opened 2023-12-06
auto activate micromamba

rolandknight:
on hermit activate, run "micromamba activate"


rolandknight:
turns out only two environment variables are required to activate the base environment. automatically.
actually running micromamba activate on trigger will not work since it can't modify parent shell.

alecthomas (approved):

PR #422, opened 2023-12-07
Fix gradle.hcl package syntax.

TadeasKriz:
In previous PR I messed up the syntax.

Note that for some reason the checks were green before merging, so maybe we're having false positives?


TadeasKriz:
@alecthomas fixed it before I opened this PR, so closing.

PR #467, opened 2024-03-26
Docker CLI

dtomasi:
Hey Alec,

This PR add docker CLI for darwin and linux starting from 20.10.

Best,
Dominik


dtomasi:
Had to remove the plugin directory from binaries, as they are only available in newer versions. But luckily, we have already a buildx plugin that can be installed separately.

PR #470, opened 2024-04-01
Add ghpc binary

casassg:
Adds ghpc binary for Google Cloud HPC tooling. 

https://github.com/GoogleCloudPlatform/hpc-toolkit/releases


adrw (approved):

staktrace (changes requested):
As discussed on Slack, this package requires all kinds of other runtime dependencies as the makefile refers to things like go and other tools. So this isn't really a good way to set up this package.

PR #526, opened 2024-09-30
feat:add kubectl argo rollouts cli

Svetixbot:
## Context
kubectl-argo-rollouts plugin allows you to manage and visualise the argo rollout and can be used as a plugin together with kubectl. 

https://argo-rollouts.readthedocs.io/en/stable/installation/


alecthomas (approved):
Thanks!

PR #530, opened 2024-10-29
feat: add android cmdline-tools, add corretto@21

stillmatic:
tested working with `openjdk-21.0.3_9`. corresponding JDK version wasn't added with corretto, added that too.

### DIFF
--- a/android-cmdline-tools.hcl
+++ b/android-cmdline-tools.hcl
@@ -1,21 +1,19 @@
 description = "Command Line Tools for Android"
-homepage    = "https://developer.android.com/tools/releases/cmdline-tools"
-binaries    = ["cmdline-tools/latest/bin/*"]
-requires    = ["jdk"]
-test        = "sdkmanager --version"
-
+homepage = "https://developer.android.com/tools/releases/cmdline-tools"
+binaries = ["cmdline-tools/latest/bin/*"]
+requires = ["jdk"]
+test = "sdkmanager --version"
 env = {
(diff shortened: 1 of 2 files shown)


android-cmdline-tools.hcl
```diff
@@ -0,0 +1,37 @@
...
+
+platform "darwin" {
+  source = "https://dl.google.com/android/repository/commandlinetools-mac-${version_}_latest.zip"
+}
+
+platform "linux" {
+  source = "https://dl.google.com/android/repository/commandlinetools-linux-${version_}_latest.zip"
+}
+
+// # this is a hack because we do not have a minor version, and hermit expects a `.` in the version
```
stillmatic, line 27: there's probably a nicer way to do this but ...

stillmatic:
hmm CI fails with 
```
 + hermit test -t android-cmdline-tools-11076708.0
fatal: app/test_cmd.go:36: env.go:548: Error: LinkageError occurred while loading main class com.android.sdklib.tool.sdkmanager.SdkManagerCli
	java.lang.UnsupportedClassVersionError: com/android/sdklib/tool/sdkmanager/SdkManagerCli has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```
the minimum JDK target is 17 and the environment only has 11. I don't see a good way in Hermit to require a particular JDK version though. (and can confirm that tests pass locally, when we explicitly use JDK 17 or 21). any idea how to make CI pass?

alecthomas (approved):

PR #543, opened 2024-12-08
kotlin-2.1.0

quad:
- Re-enable Kotlin bumps
- Add version 2.1.0
- Ignore the now missing `kotlin-dce-js` binary going forward


staktrace (approved):
I really wish the HCL schema documentation was better. Didn't know this was possible!

PR #553, opened 2025-01-08
Add kotlin-format

wsutina:
Following how ktfmt is set up https://github.com/cashapp/hermit-packages/pull/219

### DIFF
--- a/kotlin-format.hcl
+++ b/kotlin-format.hcl
@@ -1,0 +1,23 @@
+description = "A command-line tool designed to enforce consistent code formatting for Kotlin."
+binaries = ["kotlin-format"]
+source = "https://repo1.maven.org/maven2/xyz/block/kotlin-formatter/kotlin-formatter/${version}/kotlin-formatter-${version}-all.jar"
+dont-extract = true
+
+on "unpack" {
+  rename {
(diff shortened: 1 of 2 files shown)


kotlin-format.hcl
```diff
@@ -0,0 +1,23 @@
+description = "A command-line tool designed to enforce consistent code formatting for Kotlin."
```
staktrace, line 1: - [ ] You'll want the runtime-dependencies thing here too (see ktfmt.hcl) since the script executes java and without the runtime-dependencies it may run some random broken version of java

kotlin-format/kotlin-format.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+KOTLIN_FORMAT_DIR="$(dirname "${BASH_SOURCE[0]:-${(%):-%x}}")"
+exec java --add-opens java.base/java.lang=ALL-UNNAMED -Xmx512m -jar "${KOTLIN_FORMAT_DIR}/kotlin-format.jar" "$@"
```
staktrace, line 4: do you know if the `--add-opens` stuff is required? If it's not then I'd prefer to remove it

PR #557, opened 2025-01-15
Switch from runtime-dependencies to requires to get JRE

staktrace:


staktrace:
For context, this extra JRE download was causing a lot of hermit locking failures in our CI environment and it seemed a bit unnecessary to require a second JRE.

PR #567, opened 2025-02-03
fix[hadolint]: use xarch var instead of arch

kate-syberspace:
use xarch (resolves to x86_64) var instead of arch (resolves to amd64)

error message:
`fatal:hermit: https://github.com/hadolint/hadolint/releases/download/v2.12.0/hadolint-linux-amd64: download failed: 404 Not Found (404), source url: https://github.com/hadolint/hadolint/releases/download/v2.12.0/hadolint-linux-amd64`


jvmakine (approved):

alecthomas (changes requested):
Won't work as-is.

alecthomas (approved):

PR #586, opened 2025-03-10
fix: Add jlumbroso/free-disk-space to maximize disk space

wsutina:
## 📝 Description
we recently added optional support for checksumming linux/arm64 binaries, so Hermit will be backfilling those for all packages. This has been manually run locally to catch it up, which should unblock future runs.

However, this step can help reduce the change of running out of spaces in the future


wsutina:
Closing this given that this step takes 3 minute and it's a one-off issue

PR #602, opened 2025-03-21
fix(gum): strip path prefix for v0.14+

kanwren:
Since 0.14.0, `gum` has a leading directory in its artifacts; this causes errors like the following:

```
fatal:hermit: gum-0.14.5: failed to find binaries "/Users/wrenn/Library/Caches/hermit/pkg/gum-0.14.5/gum"
```

Strip this path component appropriately.


alecthomas (approved):
Awesome, thanks!

PR #638, opened 2025-07-23
feat: add ibazel

mpeyper:
Adding `ibazel`.

There is an [issue](https://github.com/bazelbuild/bazel-watcher/pull/776) with their github releases that means not every version has suitable binaries. I picked a recent-ish version that had at least `linux-amd64` `linux-arm64` and `darwin-arm64` available, but auto-versioning will not be possible for now.

### DIFF
--- a/ibazel.hcl
+++ b/ibazel.hcl
@@ -1,0 +1,29 @@
+description = "Tools for building Bazel targets when source files change."
+binaries = ["ibazel"]
+source = "https://github.com/bazelbuild/bazel-watcher/releases/download/v${version}/ibazel_${os}_${arch_}"
+
+vars = {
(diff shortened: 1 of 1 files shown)


ibazel.hcl
```diff
@@ -0,0 +1,29 @@
...
+    from = "${root}/ibazel_${os}_${arch_}"
+    to = "${root}/ibazel"
+  }
+}
+
+platform "darwin" {
+  vars = {
+    "arch_": "arm64",
+  }
+}
```
mpeyper, line 20: This prevents an error when generating the digests below (as `darwin_arm64` is not available in the github releases for the `0.26.2` version), but I'm not sure if it's actually a good approach?

PR #641, opened 2025-07-24
Add vectorscan package

binarylogic:
## Summary
- Add vectorscan package definition for hermit
- Vectorscan is a portable fork of the high-performance regular expression matching library (Hyperscan)
- Builds from source using CMake as no prebuilt binaries are available

## Test plan
- [ ] Test package installation on macOS (requires CMake, build tools)
- [ ] Test package installation on Linux (requires CMake, build tools)
- [ ] Verify library files are installed correctly
- [ ] Test with a simple C program that uses vectorscan

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/vectorscan.hcl
+++ b/vectorscan.hcl
@@ -1,0 +1,58 @@
+description = "A portable fork of the high-performance regular expression matching library."
+binaries = ["lib/libhs.so*", "lib/libhs.dylib*", "include/hs/*"]
+test = "ls ${root}/lib/libhs*"
+
+darwin {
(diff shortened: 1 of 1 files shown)


binarylogic:
Closing this PR. Vectorscan only provides source releases (no prebuilt binaries) and requires several build dependencies (boost, cmake, pcre, ragel, pkg-config). This doesn't fit well with hermit's approach of distributing prebuilt binaries.

Users needing vectorscan are better served by system package managers or building from source directly.

PR #649, opened 2025-08-09
fix(rust): auto version 'unknown variable ' bug and updates versions

erikkrieg:
Auto version CI job was failing to update rust with this error: `fatal:hermit: rust.hcl: unknown variable $version`

I believe I was able to fix this, and also added two missing versions, 1.88.0 and 1.89.0.

### DIFF
--- a/rust.hcl
+++ b/rust.hcl
@@ -6,20 +6,12 @@ env = {
   "PATH": "${HERMIT_ENV}/.hermit/rust/bin:${PATH}",
 }
 
-platform "darwin" "amd64" {
-  source = "https://static.rust-lang.org/dist/rust-${version}-x86_64-apple-darwin.tar.xz"
+darwin {
+  source = "https://static.rust-lang.org/dist/rust-${version}-${xarch}-apple-darwin.tar.xz"
 }
 
-platform "darwin" "arm64" {
-  source = "https://static.rust-lang.org/dist/rust-${version}-aarch64-apple-darwin.tar.xz"
-}
-
-platform "linux" "amd64" {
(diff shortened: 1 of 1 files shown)


rust.hcl
```diff
@@ -6,14 +6,20 @@ env = {
...
+  source = "https://static.rust-lang.org/dist/rust-${version}-aarch64-apple-darwin.tar.xz"
+}
+
+platform "linux" "amd64" {
+  source = "https://static.rust-lang.org/dist/rust-${version}-x86_64-unknown-linux-gnu.tar.xz"
+}
+
+platform "linux" "arm64" {
+  source = "https://static.rust-lang.org/dist/rust-${version}-aarch64-unknown-linux-gnu.tar.xz"
 }
```
mpeyper, line 23: I think this can be simplified down to just:

```
darwin {
  source = "https://static.rust-lang.org/dist/rust-${version}-${xarch}-apple-darwin.tar.xz"
}

linux {
  source = "https://static.rust-lang.org/dist/rust-${version}-${xarch}-unknown-linux-gnu.tar.xz"
}
```

It's unclear why to me at this stage, but the error only seems to trigger when the `sha256-source` field tries to use the `version` variable.

mpeyper (approved):
Thanks for this. 

FWIW, `source + ".sha256"` is already one of the [standard locations hermit looks for sha256 sources](https://github.com/cashapp/hermit/blob/d82d9091e75d2668fb601e62e205ec0e8cb156b4/manifest/digest/digest.go#L180-L182), so removing the explicit `sha256-source` should be safe to do in this case.

PR #658, opened 2025-09-28
fix: add arm64 for hadolint

nickajacks1:
Starting in 2.13.1, hadolint includes arm64 builds and renamed darwin to "macos".


alecthomas (approved):
Thanks!

PR #691, opened 2025-11-19
Fix podman package for Darwin/Linux platform compatibility

Gussy:
Installing podman on macOS currently fails as the release package structure changed.

```text
❯ hermit install podman
fatal:hermit: podman-5.7.0: no binaries found matching "/Users/.../Library/Caches/hermit/pkg/podman-5.7.0/podman" - this may indicate a corrupted or misconfigured package, try removing /Users/.../Library/Caches/hermit/pkg/podman-5.7.0 and hermit will reinstall it (may require sudo)
```

This fixes the issue for 5.7.0, although previous versions were untested.

### DIFF
--- a/podman.hcl
+++ b/podman.hcl
@@ -1,22 +1,17 @@
 description = "A tool for managing OCI containers and pods."
 sha256-source = "https://github.com/containers/podman/releases/download/v${version}/shasums"
-binaries = ["podman"]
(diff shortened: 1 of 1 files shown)


alecthomas:
I think I'll close this one out as it's been so long I expect there will be bitrot. Apologies for not getting to this earlier.

PR #742, opened 2026-03-25
zellij: fix sha256-source url and wrong checksum

alicul:
Root cause: the linux package block was trying to download the zellij-no-web-... tarball, but its sha256-source was mistakenly pointing to the standard zellij-... .sha256sum file. This caused hermit to populate the sha256sums block with the checksum of the standard version (d7239...) instead of the no-web version (458b0...).

Started failing after: https://github.com/cashapp/hermit-packages/commit/b4ca316e5742acbc778245a43812a4e7df787e1f

### DIFF
--- a/zellij.hcl
+++ b/zellij.hcl
@@ -4,7 +4,7 @@ test = "zellij --version"
 
 linux {
   source = "https://github.com/zellij-org/zellij/releases/download/v${version}/zellij-no-web-${xarch}-unknown-linux-musl.tar.gz"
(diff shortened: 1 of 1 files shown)


aliculPix4D:
@alecthomas friendly ping..

aliculPix4D:
@alecthomas Is there any reason why some PRs like this one never get reviewed/merged? If necessary, I would be more than willing to address any review comments. Thanks.

odormond (approved):

alecthomas (approved):

PR #748, opened 2026-04-20
Make platform blocks explicit for opa and yajsv

Renjuju:
## Summary
- **opa.hcl**: Change `linux {}` and `darwin {}` to `platform "linux" "amd64" {}` and `platform "darwin" "amd64" {}` in the v0.30.1 block so hermit doesn't try to install amd64 binaries on arm64 machines. Newer versions (0.38.1+) already use `${arch}` and support arm64 natively.
- **yajsv.hcl**: Replace top-level source (which matched all platforms including arm64) with explicit `platform "linux" "amd64" {}` block since no linux-arm64 binary exists. Darwin keeps the arch-unqualified block (works via Rosetta on arm64 macs).

Same pattern as #747 (codecov).

### DIFF
--- a/opa.hcl
+++ b/opa.hcl
@@ -3,7 +3,7 @@ binaries = ["opa"]
 test = "opa version"
 
 version "0.30.1" {
-  linux {
+  platform "linux" "amd64" {
     source = "https://github.com/open-policy-agent/opa/releases/download/v${version}/opa_linux_amd64"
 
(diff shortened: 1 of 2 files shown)


Renjuju:
Closing as part of a stale-PR cleanup sweep — no activity since before May. Reopen if this is still wanted; the branch is left intact.

PR #759, opened 2026-05-29
Add OpenBao

dtomasi:
Adds OpenBao from https://github.com/openbao/openbao

> OpenBao is an open source, community-driven secrets manager and fork of Vault managed by the Linux Foundation's OpenSSF


alecthomas:
Thanks and apologies for the delay.

PR #762, opened 2026-06-11
Update ktfmt source and version references

kcolford:
At version 0.55 and up, the jar file has been renamed to `ktfmt-${version}-with-dependencies.jar` and so it's impossible to download a newer version without this


alecthomas (changes requested):
We don't allow older versions to be deleted. The correct fix is to have two version blocks:

- The old one with the old source and no `auto-version`
- The new one from 0.55 and up with the global source and `auto-version`

PR #767, opened 2026-06-23
Use x86_64 for instead of amd64 for the arch

staktrace:


staktrace:
I tested this using a local docker ubuntu image running with `--platform linux/amd64`. Installing this package failed without this change and passed with the change.

ChristianHamm:
Thank you very much!

PR #760, opened 2026-06-07
OCM: Align with OCM's new canonical release strategy starting from v0.8.0

dtomasi:
Starting from v0.8.0 OCM uses combined/canonical release strategy, where CLI is no longer available under its own git tag with prefix.

This fix migrates to the new layout by maintaining backward compatibility to previous versions.

PR #778, opened 2026-07-21
Uses new version format for OCM

ViMaSter:
Starting with 0.8.0, OCM changed their version syntax: https://github.com/open-component-model/open-component-model/releases?page=2#release-v0.8.0

With this change, new versions follow the new format, old version remain supported, and all currently available version are added once. The auto-version action should then be able to pick up future releases again.


ViMaSter:
Duplicate of #760.

PR #785, opened 2026-07-27
Add pgsql

iilyak:
Add PostgreSQL which is a powerful, open source object-relational database system with over 35 years of active development that has earned it a strong reputation for reliability, feature robustness, and performance.


alecthomas (commented):
There don't seem to be Linux binaries?

alecthomas:
In general unless there are really good reasons, we don't include packages that are only for a single platform.

PR #775, opened 2026-07-13
fix broken `cargo-nextest.hcl` update URL

OmarTawfik:
I noticed the `Auto-version` CI has been failing to update this package:

- https://github.com/cashapp/hermit-packages/actions/runs/29222227077/job/86729518406

```log
warn:cargo-nextest: Retrying. Failed to download any of https://github.com/nextest-rs/nextest/releases/download/cargo-nextest-cargo-nextest-0.9.140/cargo-nextest-cargo-nextest-0.9.140-aarch64-unknown-linux-gnu.tar.gz on attempt 1/3: download failed: 404 Not Found (404), source url: https://github.com/nextest-rs/nextest/releases/download/cargo-nextest-cargo-nextest-0.9.140/cargo-nextest-cargo-nextest-0.9.140-aarch64-unknown-linux-gnu.tar.gz
warn:cargo-nextest: Retrying. Failed to download any of https://github.com/nextest-rs/nextest/releases/download/cargo-nextest-cargo-nextest-0.9.140/cargo-nextest-cargo-nextest-0.9.140-aarch64-unknown-linux-gnu.tar.gz on attempt 2/3: download failed: 404 Not Found (404), source url: https://github.com/nextest-rs/nextest/releases/download/cargo-nextest-cargo-nextest-0.9.140/cargo-nextest-cargo-nextest-0.9.140-aarch64-unknown-linux-gnu.tar.gz
warn:cargo-nextest: Failed to download any of https://github.com/nextest-rs/nextest/releases/download/cargo-nextest-cargo-nextest-0.9.140/cargo-nextest-cargo-nextest-0.9.140-aarch64-unknown-linux-gnu.tar.gz on attempt 3/3: download failed: 404 Not Found (404), source url: https://github.com/nextest-rs/nextest/releases/download/cargo-nextest-cargo-nextest-0.9.140/cargo-nextest-cargo-nextest-0.9.140-aarch64-unknown-linux-gnu.tar.gz
[...]

PR #789, opened 2026-08-05
cargo-nextest: update to 0.9.143

steobrien:
Auto-version has been stuck at 0.9.68 because nextest tags releases as cargo-nextest-<version>, so this adds a version-pattern to match those tags, plus the latest version and its digests.


teofr:
Just pointing out there's already a PR that I think fixes the underlying problem https://github.com/cashapp/hermit-packages/pull/775

alecthomas (approved):

alecthomas:
Thanks! This is the correct fix, I'll close out the other one.

PR #791, opened 2026-08-12
Add OutageDeck package

koko3tallah:
Adds [OutageDeck CLI](https://github.com/outagedeck/cli) to Hermit's public package registry. It checks live status for cloud and SaaS dependencies from their official status feeds, without requiring an API key.

The manifest installs the current macOS or Linux release archive on AMD64 and ARM64, exposes the `outagedeck` binary, and follows future GitHub releases automatically.

I'm Kerolos, founder of OutageDeck, and maintain the release assets used here.

## Test plan

- [x] `hermit validate "file://$PWD"`
- [x] Registry search resolves `outagedeck@0.1.3`
- [x] Native macOS ARM64 install returns `0.1.3`
- [x] Native macOS ARM64 live provider check succeeds
- [x] Linux AMD64 install returns `0.1.3` under local architecture emulation
- [x] Release archive digests match the published GitHub assets

### DIFF
--- a/outagedeck.hcl
+++ b/outagedeck.hcl
@@ -1,0 +1,19 @@
+description = "Check live cloud and SaaS provider status from the terminal"
+homepage = "https://outagedeck.com"
+repository = "https://github.com/outagedeck/cli"
+binaries = ["outagedeck"]
+test = "outagedeck --version"
+source = "https://github.com/outagedeck/cli/releases/download/v${version}/outagedeck_${version}_${os}_${arch}.tar.gz"
+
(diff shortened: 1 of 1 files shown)


alecthomas (approved):

koko3tallah:
The package description I contributed says `live`, but OutageDeck ingests vendor feeds every ten minutes. Could you replace it with:

`Check current cloud and SaaS provider status from the terminal`

This is a copy-only correction. The manifest behavior, release source, and hashes do not change.

PR #792, opened 2026-08-17
update oapi-codegen description and add new version binaries

karstenkoehler:
Updates oapi-codegen to the current module path and adds more versions. I'd love to use version 2.8.0, but it seems like automatic workflow runs for https://github.com/cashapp/hermit-build have been disabled. Would it be possible to activate this again, so that it builds the latest version of oapi-codegen? I can then include version 2.8.0 into this PR.


alecthomas (approved):

alecthomas:
Re-enabled - GHA automatically disables them unfortunately.

PR #793, opened 2026-08-17
mdbook: add support for macOS ARM

marco-m-pix4d:
It was a breeze!
Thanks for Hermit!

How I tested

I had a repo that contained already a `bin/hermit.hcl` and a `bin/mdbook` that was added, from Linux, with `hermit install mdbook@0.4`.

I edited that `bin/hermit.hcl` to point to my checked-out `hermit-packages` repo with the current branch:
```
sources = ["file:///Users/banana/src/wo/hermit-packages"]
```

I cd out and inside again of the target repo, on macOS ARM, with a shell configured with Hermit shell integration.

Ran `mdbook --version`. It got downloaded and installed by Hermit.

Added checksums with
```
hermit manifest add-digests mdbook.hcl
```

A few did not exist, but according to https://github.com/cashapp/hermit-packages/pull/698#discussion_r2624335076, nothing needs to be done.

### DIFF
--- a/mdbook.hcl
+++ b/mdbook.hcl
@@ -10,6 +10,10 @@ platform "darwin" "amd64" {
   source = "https://github.com/rust-lang/mdBook/releases/download/v${version}/mdbook-v${version}-${xarch}-apple-darwin.tar.gz"
 }
 
+platform "darwin" "arm64" {
+  source = "https://github.com/rust-lang/mdBook/releases/download/v${version}/mdbook-v${version}-${xarch}-apple-darwin.tar.gz"
+}
+
 version "0.4.43" "0.4.44" "0.4.45" "0.4.47" "0.4.48" "0.4.49" "0.4.50" "0.4.51"
         "0.4.52" "0.5.0-beta.1" "0.5.0-beta.2" "0.5.0" "0.5.1" "0.5.2" "0.5.3" "0.5.4" {
   auto-version {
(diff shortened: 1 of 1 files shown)


mdbook.hcl
```diff
@@ -10,6 +10,10 @@ platform "darwin" "amd64" {
   source = "https://github.com/rust-lang/mdBook/releases/download/v${version}/mdbook-v${version}-${xarch}-apple-darwin.tar.gz"
 }
 
+platform "darwin" "arm64" {
+  source = "https://github.com/rust-lang/mdBook/releases/download/v${version}/mdbook-v${version}-${xarch}-apple-darwin.tar.gz"
```
aliculPix4D, line 14: I expected to see some sha256sums being added?

Also, are we sure that each version in the list below (L17) has its own corresponding arm64 Darwin release?

mdbook.hcl
```diff
@@ -10,6 +10,10 @@ platform "darwin" "amd64" {
   source = "https://github.com/rust-lang/mdBook/releases/download/v${version}/mdbook-v${version}-${xarch}-apple-darwin.tar.gz"
 }
 
+platform "darwin" "arm64" {
```
aliculPix4D, line 13: why duplicate?

why not just remove `amd64` on L9 and support both amd64 and arm64? The original "amd64" was there to limit Darwin to Intel. See how linux block already handles both architectures.

So the only "issue" I see with that is that probably some version will be lacking arm64 e.g. `0.4.43` so we can create a separate block only for that version to limit it to amd64 only.

mdbook.hcl
marco-m-pix4d: Added checksums.
See https://github.com/cashapp/hermit-packages/pull/698 for the "are we sure" question, as I understand it.

